PR #8, opened 2026-09-03
fix(ci): unbreak Validate and Security, and stop a 429 masking a dead link

napetrov:
## What this changes

`main` is red on two jobs and every open pull request inherits the first of them. This
fixes both, plus the dependabot config gap that makes CodeQL unpassable on PRs #5 and #6.

**Validate / Link check.** `skills/vllm-xpu-run/SKILL.md` points at
`docs.vllm.ai/en/latest/getting_started/xpu-installation.html`. docs.vllm.ai dropped the
`.html` URL scheme, so it now redirects to a path that 404s. The link really is dead — but
the file is an import pinned to `intel/gpu-ai-skills`, and editing it here would move the
failure one step down to `sync_external.py --check`, which byte-compares the copy against
its pin. Check 8 already resolves this exact tension by warning on an imported body rather
than failing it; the link check now does the same, and the warning says where the fix has
to land. A link this repository wrote still fails, and so does a pinned commit that has
gone missing — that one is our claim, not upstream's.

**Link check, second defect.** The same tree passed and failed three seconds apart this
morning: PR #3's run got `HTTP 429` from docs.vllm.ai and went green, PR #4's got the real
`404` and went red.

```
04:48:39  WARN ... xpu-installation.html was not reachable (HTTP 429)   ← pass
04:48:42  FAIL ... xpu-installation.html is gone (HTTP 404)             ← fail
```

Rate limiting maps to `unknown`, which warns. That is right for someone else's outage and
wrong here, because 429 is the one answer a host gives *instead of* the real one — eight
workers over ~100 URLs concentrated on a few hosts is enough to trigger it. It is now
retried with backoff, honouring `Retry-After` when it is in seconds, before it may become
a warning. A throttled run can no longer report a dead link as reachable.

**Validate step order.** `Imported skills match their pinned upstream` runs after the link
check with no `if: always()`, so it has never executed on a commit with a dead link — the
state of the imports was unknown for as long as `main` was red. It now always runs, and it
passes.

**Security / zizmor** — 5 high, all from the Pages workflow added in #7:

| finding | fix |
|---|---|
| `pages: write`, `id-token: write` at workflow level | moved to the `deploy` job. `build` is where third-party code runs (`npm ci`, `npm run build`) and is the job that should not hold a token that can publish the site or mint an OIDC identity |
| `actions/configure-pages@v5` unpinned | `@983d7736` (v5.0.0) |
| `actions/upload-pages-artifact@v3` unpinned | `@56afc609` (v3.0.1) |
| `actions/deploy-pages@v5` unpinned | `@368f8252` (v5.0.1) |

Each SHA is the one behind the tag that was already resolving, so nothing about what runs
changes. Newer majors exist (configure-pages v6, upload-pages-artifact v5) and are
deliberately left to dependabot, which will propose them with a CI run attached.

**CodeQL on dependabot PRs #5 and #6.** `codeql-action/init` and `codeql-action/analyze`
are two entry points into one action, and it refuses a version split between its halves:

```
PR #5: Loaded a configuration file for version '4.37.9', but running version '4.36.2'
PR #6: Loaded a configuration file for version '4.36.2', but running version '4.37.9'
```

Ungrouped, dependabot opens one pull request per path, so each bumps half the pair and
neither can pass its own CI. `.github/dependabot.yml` now groups `github/codeql-action*`.
**#5 and #6 need closing** so the grouped pull request can replace them.

## Verified locally

All four gates, plus both linters from the same digest-pinned images `security.yml` uses:

```
validate_skills.py                 OK 33 skill(s)
validate_skills.py --check-links   link check: 99 URL(s), 1 gone, 1 of them only in
                                   imported bodies (warned)          exit 0
run_evals.py --validate            OK 10 eval file(s), 30 case(s)
sync_external.py --check           OK, every import matches its pin
actionlint                         clean
zizmor                             No findings to report (12 suppressed)   was 5 high
```

The imported-body path was unit-checked in both directions: a dead link in a body this
repository wrote still errors, and a rewritten `external-commit` still errors. The 429
path was checked against a stubbed opener — `429 → 429 → … → warn`, and critically
`429 → 404 → dead`, which is the case that used to buy a green run.

## Not in this change

The three stale docs.vllm.ai URLs themselves. They live in imported bodies, so they need a
pull request against `intel/gpu-ai-skills` and then a pin move here, which is a follow-up
PR. Two of the three are worse than the one CI caught:

| file | link | actually resolves to |
|---|---|---|
| `vllm-xpu-run/SKILL.md` | `getting_started/xpu-installation.html` | 404 |
| `vllm-xpu-profile/SKILL.md` | `contributing/profiling/profiling_index.html` | 200 — `…/contributing/`, the *parent* page |
| `vllm-xpu-bench/SKILL.md` | same + `#offline-batched-inference-benchmarks` | 200 — same parent page |

Read-the-Docs fuzzy-redirects the latter two, so they answer 200 while pointing an agent at
the wrong page. No status-based link check can see that; it took reading the redirect
chain. Correct targets are
`getting_started/installation/gpu.html?device=xpu` and `contributing/profiling/`.

One more thing this change cannot fix: #7 merged with Validate, Security and DCO all
failing, which is how `main` got here. Worth requiring `validate`, `install`, `zizmor`,
`actionlint` and `signed-off` as status checks on `main` — otherwise every tool above is
advisory.

## Checklist

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -48,8 +48,10 @@ jobs:
 
       # Its own step because it is the one check that depends on someone else's
       # server. A 404 or 410 fails it — that is a pointer an agent would follow
-      # into nothing. A timeout, a 5xx, or rate limiting only warns, so an outage
-      # somewhere on the internet cannot hold a contributor's pull request.
+      # into nothing — unless the link came from an imported body, where it warns
+      # instead, because the repair has to land upstream and arrive through a moved
+      # pin. A timeout, a 5xx, or rate limiting only warns, so an outage somewhere on
+      # the internet cannot hold a contributor's pull request.
       - name: Link check
         run: python3 tools/validate_skills.py --check-links
 
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -203,6 +203,15 @@ written, and that is what the pin records. Keep the imported text as it was — 
 it means existing measurements of the original no longer describe the file here. The one
 edit the generator does make is to repository-relative command paths, which would
 otherwise point at upstream's layout and resolve to nothing once the skill is installed.
+
+That is also why two checks report an imported body as a warning where they would fail a
+skill written here: an unmentioned file, and a link that has gone 404. Both name a real
+defect and neither can be fixed in this repository — editing the body breaks the
+byte-compare in `sync_external.py --check`, which is the thing that proves the copy is
+still what was reviewed. The route is a pull request upstream, then move
+`external-commit` and re-run `--write`. A warning that outlives a release is worth
+raising with the upstream maintainer rather than living with; if upstream will not take
+the fix, the pin is the wrong pin.
 Every file it touches is listed in `.source.json` under `modified-files` and carries a
 one-line notice saying so, and the validator fails if either is missing.
 
--- a/tools/validate_skills.py
+++ b/tools/validate_skills.py
@@ -63,7 +63,10 @@ Everything above is offline. `--check-links` adds the guide's "link check": it
 resolves the http(s) links in each SKILL.md, and for an imported skill the upstream
 repository and its pinned commit, so a moved or deleted target is caught before a
 reader follows a dead reference. It is opt-in and a separate CI step because it is
-the one check whose result depends on someone else's server being up.
+the one check whose result depends on someone else's server being up. A dead link in
+an imported body warns rather than fails, for the same reason 8 does: the fix belongs
+upstream, and an edited import no longer matches its pin. The pinned commit itself
+still fails, because that one is this repository's own claim rather than upstream's.
 
 Needs Python 3.11 or newer for tomllib. Nothing else outside the standard library.
 """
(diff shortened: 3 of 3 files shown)


napetrov:
Superseded — split into five single-purpose pull requests, since the combined change mixed a policy decision, a determinism fix, a step-ordering bug, a workflow-permissions audit and a dependabot config gap under one review.

| # | scope | base |
|---|---|---|
| #9 | a dead link in an imported body warns instead of failing — **unblocks `main`** | `main` |
| #10 | retry a 429 in the link check instead of accepting it | #9 |
| #11 | run the pinned-upstream check even when the link check failed | #9 |
| #12 | scope the Pages token to the deploy job and pin its actions — **unblocks Security** | #9 |
| #13 | group the codeql-action pins so both halves move together | #9 |

#10 through #13 are stacked on #9 only so their Validate runs are green while `main` is red; GitHub retargets them to `main` when #9 merges. All five merge into `main` cleanly in any order — verified by test-merging the set, then running the full gate plus zizmor and actionlint on the result.

Closing this one.

PR #18, opened 2026-09-09
Readme restructure

napetrov:
## What this changes

Updating readme and making it more generalized


xaleryb (commented):
## Review

No blockers. Three notes.

**The Quickstart now makes a third-party CLI the default install command. Two lines worth
adding next to it:**

- Telemetry — `src/telemetry.ts:1-2` posts to `add-skill.vercel.sh/t` and queries `/audit`;
  `track()` fires on install (`add.ts:1833`), on unless `DISABLE_TELEMETRY`/`DO_NOT_TRACK`,
  and because this repo is public the owner/repo/skill identifiers are included. Worth saying
  where the command is given.
- A version floor. `vercel-labs/skills#1138`: a project-local install into a repo with no
  pre-existing `.claude/` wrote the skill to `.agents/skills/` and silently skipped the
  `.claude/skills/<name>` symlinks — Claude Code never loaded it, with no warning and a
  summary that read as success. Fixed by `#1598`, first shipped in v1.5.16. That is exactly
  this Quickstart's default case (project-scoped, `--agent claude-code`, fresh repo), and
  plain `npx skills` reuses whatever CLI is already cached. Either
  `npx skills@latest add intel/skills` or an explicit "v1.5.16 or newer" covers it.

**The one-task rule is now in both documents** — `## Trust and validation` in the README and
CONTRIBUTING (`:15`, `:221`, `:316`). Fine if that is deliberate; flagging it because the two
copies will drift.

FYI on the substance: the catalog doesn't fully meet that rule yet. Seven of the ten skills
written here have no task — 15 task directories, and every `task.toml` names
`dpnp-quickstart`, `mkl-extension-advisor` or `onetbb-quickstart` in `[metadata] skill`.
Nothing to change in this PR; suggest backfilling the missing tasks in a follow-up so the
statement holds.

**Breaks on merge:** moving the merge bar out of the README makes `MAINTAINERS.md:4`, `:58`
and `:22` wrong. Happy to send these three as a follow-up if you want to keep this PR to four
files.

README.md
```diff
@@ -1,284 +1,195 @@
...
+## Quickstart
 
-You need a GitHub account, working knowledge of the thing your skill covers, and an editor.
-No Intel hardware, no benchmark data and no credentials — nothing that a fork cannot get to.
-Field-by-field reference: [CONTRIBUTING.md](CONTRIBUTING.md).
+Install with the standard skills CLI, which reads this repository directly:
 
-There are two ways in, and they ask for different things.
+```bash
+npx skills add intel/skills
```
xaleryb, line 23: Version floor, as a one-word change. Plain `npx skills` reuses whatever CLI version is already cached, and the silently-skipped `.claude/skills/<name>` symlinks (`vercel-labs/skills#1138`) are only fixed from v1.5.16 onwards.

```suggestion
npx skills@latest add intel/skills
```

If you want it strict, the same applies to the three variants below - your call. In practice this first invocation refreshes the cache for them.

README.md
```diff
@@ -1,284 +1,195 @@
...
-mkdir -p skills/your-skill-name          # kebab-case, ≤ 64 characters
-cp templates/SKILL.md skills/your-skill-name/SKILL.md
+# for every project on the machine rather than this one
+npx skills add intel/skills --skill linux-perf --agent claude-code --global
 ```
 
-The directory name is the skill's identity — it must equal `name:` in the frontmatter.
+Installs are project-scoped by default — `./.claude/skills/<name>` for Claude Code, and the
+equivalent directory for each other agent you name. `--global` writes the per-user location
+instead. Check your agent's own documentation for the directory it reads.
```
xaleryb, line 41: Telemetry note from my earlier comment, as a concrete wording. Checked against the CLI: `isEnabled()` in `src/telemetry.ts` is `!DISABLE_TELEMETRY && !DO_NOT_TRACK`, so any non-empty value turns it off.

```suggestion
instead. Check your agent's own documentation for the directory it reads.

The CLI is a third-party tool (`vercel-labs/skills`, MIT) and reports installs to
`add-skill.vercel.sh` by default; `DISABLE_TELEMETRY=1` or `DO_NOT_TRACK=1` turns that off.
The offline path below sends nothing.
```

xaleryb (commented):
The two Quickstart items from my earlier comment, as suggestions you can commit directly. Still no blockers.

xaleryb (approved):

README.md
napetrov: We can keep simple commands here and update details in docs

README.md
napetrov: Worth just focus or notes, not main readme.md

PR #19, opened 2026-09-09
ci(security): scan every skill with SkillSpector

napetrov:
Adds a fourth CI gate: every skill in `skills/` is scanned with
[NVIDIA SkillSpector](https://github.com/NVIDIA/SkillSpector) (71 patterns / 17 categories,
AST pass over shipped scripts, YARA signatures, OSV.dev lookup for named dependencies).

Draft because one decision is the maintainers' to make, not mine — see **Open questions**.

## What is here

| file | why |
| --- | --- |
| `.github/workflows/skillspector.yml` | PR + push-to-main + weekly cron. Keyless (`--no-llm`), so a fork gets the same answer as a branch. SkillSpector ships no action and no container image, so it is `pip install`ed from a **full 40-char commit pin** — nothing third-party is handed a token and no allowed-actions policy needs a new entry. Plus one step that re-reads an import whose pin this PR moved. |
| `tools/scan_skills.py` | SkillSpector scans one skill at a time; this repo has 33. Runs them concurrently, gives each the baseline rules scoped to it, writes the table to the step summary, exits 1 on a skill above the threshold its origin is held to, on a scan that did not complete, or on a merged SARIF missing a skill that scanned. Adds no dependency of its own: it reads YAML through `pyyaml`, which SkillSpector already requires. |
| `.skillspector-baseline.yaml` | The reviewed suppressions, with a reason each and a `skills:` list naming the skills each applies to. |
| `CONTRIBUTING.md` | How to run it locally and what to do with a finding. |

## Two thresholds, because the two kinds of skill can act on a finding differently

A skill written here can be fixed in the pull request that reports the problem. An imported
skill is upstream's text, kept byte-for-byte because `sync_external.py --check` compares it
to the pinned commit — editing it here would break the thing that makes an import worth
having, so its repair has to land upstream and arrive through a moved pin. That is the same
asymmetry that already makes the mentions check and the link check warnings in an imported
body.

| origin | fails above | why that number |
| --- | --- | --- |
| written here (10 skills) | **20** | SkillSpector's LOW/`SAFE` band — where all ten already sit |
| imported (23 skills) | **50** | the boundary above which the scanner itself says `DO_NOT_INSTALL` |

Findings under an import's threshold are printed, not failed, and the table marks which
body each is in. When a PR *moves* a pin, one extra step re-reads those imports at the
stricter threshold and prints the result — new upstream text is the moment to read its
findings again. Reported, not blocking: failing it would block the bump on work that cannot
be done in that PR. Locally that is `tools/scan_skills.py --strict-imports <skill>`.

Neither number was picked here; both are `nodes/report.py`'s own band edges (`0-20` LOW,
`21-50` MEDIUM, `51-80` HIGH, `81+` CRITICAL).

## Baseline: rules only, never fingerprints

`skillspector baseline` generates fingerprint suppressions bound to the exact bytes they
were made from. That is the wrong instrument for a catalog under active edit: every
reworded paragraph would reactivate the finding it accepted, and the gate would go red on
content changes that have nothing to do with security. So the file carries only
drift-tolerant `rules`, split into two kinds:

- **POLICY** — the pattern is inherent to what this catalog documents and is not going to
  be fixed: render-node passthrough (`--device /dev/dri`), host IPC for a multi-process
  inference server (`--ipc=host`), host networking, the documented `sudo` protocol, curl
  against a loopback endpoint the skill just started, first-party tools shelling out to
  `perf`/`addr2line`/`docker` with list argv, and a handful of English sentences a regex
  reads as instructions (`swap instructions` → memory poisoning, `display rules` → system
  prompt leakage, an `<!-- (C) Intel … -->` header → hidden instructions).
- **TRACKED** — real, not fixed here. Currently only `xpu-port`'s
  `exec(compile(code, …))`, which is that tool's documented purpose and already carries a
  Bandit justification at the call site.

`--privileged`, `nsenter`, `systemctl enable` and the `pip install git+https://` are
deliberately **not** suppressed. They are the findings worth re-reading each time they move.

### Each rule names the skills it applies to

SkillSpector's rule schema is `id`, `path`, `message`, `reason`, and `path` matches a
finding's file *within* a skill (`SKILL.md`, `scripts/foo.py`) — nothing in it can say which
skill an entry is for. One shared file therefore reaches all 33: the entry accepting `exec()`
in `xpu-port`'s verifier would also accept it in a skill nobody has written yet. Measured with
a probe skill shipping one `subprocess` call: **0, nothing active** under a blanket `AST4`
entry, **9 with the finding active** once `AST4` names the two skills it was written for.

So the baseline carries a `skills:` list this repository defines and `tools/scan_skills.py`
enforces, by writing each skill a baseline holding only the rules that name it. Two native
alternatives were tried and neither works: the scanner refuses a shared baseline for a
recursive multi-skill scan, and a per-skill file inside `skills/<name>/` fails
`sync_external.py --check` on any imported skill, which is where nearly every finding is. A
scope matching no skill fails the run rather than reading as cover. 30 of the 33 entries are
scoped, taking each skill from 30 rules to 6–13; `RP1`, `PE5` and `LP3` stay unscoped, each a
statement that the rule is inapplicable to this catalog as a whole. The catalog's result is
unchanged by all of it — 17 active, worst 35, every per-skill score identical — which is the
point: it closes what the baseline promises about future skills without moving today's answer.

### Why `RP1` is POLICY (corrected from an earlier revision of this description)

I first read `RP1`'s 36 hits as unpinned image references worth fixing across 10 skills.
Reading them line by line says otherwise, on three counts:

- Its subject is an **MCP server** rug-pull, and no skill in this catalog declares an MCP
  server. What it matched is `docker run`, `docker pull` and `pip install` lines.
- Its pin check reads the matched text and the remainder of **that one line**
  (`nodes/analyzers/mcp_rug_pull.py`). So a `docker run --rm \` whose image tag sits on the
  next line reads as unpinned — 34 of the 36 hits are of that kind, plus a prose table cell
  (`| docker run --gpus all |`), a Python comment, and `xpu-container-run/SKILL.md:3`, which
  is the frontmatter `description:` line.
- Pinning is the wrong advice here anyway. The current tag of `intel/vllm`,
  `intel/sglang-dev` and `intel/oneapi` is the track Intel supports; a skill that pinned one
  in prose would send its reader to an image without the fix they came for. Where a version
  does change the answer the skills already say to pin one (`intel/vllm:<version>-xpu`), and
  a measurement that has to reproduce records the resolved image in `perf/`.

## Where a finding shows up (not only in the job log)

Three surfaces, each chosen so a fork PR still gets an answer:

| surface | what it carries | fork PR? |
| --- | --- | --- |
| **step summary** | the 33-row table + tally, and for every skill the diff **touched**, each finding in full — matched text, what the rule says, what it wants done (`--detail`) | yes |
| **annotations on the diff** | `::error`/`::warning` per finding on its own file and line, so it renders in **Files changed** (`--annotate`). GitHub renders 10 per level per step, so they are ordered over-threshold first, then touched skills, then by score, and a `::notice` says how many were dropped | yes |
| **code scanning alerts** | one alert per active finding, PR shows the ones its diff introduced, Security tab keeps the history across commits (`--sarif` → `upload-sarif`, same digest-pinned action `codeql.yml` uses, `category: skillspector`) | no — read-only token; the run says so in the summary instead |

Which skills a PR touched comes from `git diff --name-only <base> HEAD -- skills/`, names filtered
to kebab-case before they reach a shell.

**Suppressed findings are deliberately not uploaded.** SkillSpector writes each one out with
its baseline `reason` in `suppressions[].justification`, which reads like it should land as a
dismissed alert holding that sentence. Measured against the API, it does not — all 214 results
came back **open**, with `kind: external` and `kind: inSource` alike. Uploading them would put
197 alerts nobody will act on in front of the 17 that want reading. So only the active ones go
up; the reasons stay in the baseline file and the counts stay in the table's `suppressed`
column.

Verified end-to-end on a same-repo PR in my fork (a PR against this repo cannot exercise the
upload): `Successfully uploaded results`, analysis registered, **17 alerts**, every path
resolving to a real file. SARIF costs a second pass over every skill — the CLI writes one
format per run, and the gate reads the score out of the JSON — so it is only requested when
there is somewhere to send it: 1m28s for a fork PR, 3m27s for a run that uploads.

## Measured on this tree (SkillSpector v2.11.2, static, 8 jobs, 35s wall / 1m28s in CI)

Unsuppressed: **12 of 33 skills score `DO_NOT_INSTALL`** (7 of them CRITICAL, worst 100/100),
214 findings.

With the baseline: **26 of 33 skills at 0, worst 35, 17 findings active.** Gate passes on
`main` today. Every non-zero score is an imported skill; all ten written here are at 0.

The table prints the score's band *and* the worst active severity
(`risk_assessment.max_issue_severity`), because the two answer different questions: six skills
pass with an active HIGH, and `xpu-profile-unitrace` sits in the LOW band while carrying an
unsuppressed one. Reading LOW as safety is the mistake a gate should not invite. The scanner's
own `recommendation` field is deliberately not used — it is `CAUTION` for all 33 skills,
including ones scoring 0 with nothing active, because its reference resolver counts filenames
mentioned in prose as unresolved.

What stays active — the whole list, nothing hidden:

| rule | sev | n | what it is | skills |
| --- | --- | --- | --- | --- |
| `PE5` | HIGH | 14 | `--privileged` (13), `nsenter` (1, in a sentence telling you not to use it) | 5 imports |
| `RA2` | MEDIUM | 2 | `systemctl enable --now docker` | `xpu-system-setup` |
| `YR1` | HIGH | 1 | `pip install git+https://…/transformers.git` | `torch-xpu-run` |

`YR1` is the one that deserves the weight. `torch-xpu-run/SKILL.md:230` offers "install from
source if needed" as the recovery path for an unrecognized model architecture, which tells
the agent to install unreviewed `main` of a third-party project into the user's environment
— a HIGH YARA match (`agent_skill_remote_bootstrap_execution`, confidence 0.85), and unlike
a moving image tag on an Intel-published image it is arbitrary upstream HEAD. The fix is one
line and belongs upstream in `intel/gpu-ai-skills`: `pip install -U transformers` for the
released wheel, and a tagged install (`…/transformers.git@v4.57.0`) only where a model card
demands a pre-release. Filed upstream as
[intel/gpu-ai-skills#13](https://github.com/intel/gpu-ai-skills/issues/13) with that diff. It
is left **unsuppressed** on purpose, so the report keeps naming it and every pin bump reads it
again — a torch-xpu-run bump is exactly what the new step surfaces.

Also verified locally with the repo's own pinned linters: `actionlint` clean, `zizmor
--min-severity=medium --min-confidence=medium` clean, `validate_skills.py`,
`sync_external.py --check` and `run_evals.py --validate` still pass.

## Open questions

1. **`PE5`/`--privileged`: suppress as POLICY, or fix upstream?** All 13 are in imported
   bodies (`xpu-container-run` documents it as a last resort with its own section;
   `sglang-xpu-run` has three unconditional launch blocks that use it, `SKILL.md:65,122,213`).
   Under the split they pass — worst is 35 against 50 — so nothing is blocked either way. I
   would leave them visible and take the three unconditional ones upstream.
2. **Should a pin bump block?** Today the re-read step reports. Making it blocking would
   turn "bump the pin" into "bump the pin and triage 13 pre-existing findings", which is why
   it does not.

Two smaller notes: dependabot will not move the scanner pin (it watches `github-actions`,
and this repo has no Python manifest), so the weekly run is what makes a stale pin and
newly-learned OSV data visible — bumping the pin should be its own PR, since the baseline
suppresses by rule id and a new version can report rules nobody here has triaged. And the
scan is static-only on purpose; the table says `scan mode: static only` so a low score is
never mistaken for a clean full scan.


.skillspector-baseline.yaml
```diff
@@ -0,0 +1,176 @@
...
+  - id: "RP1"
+    reason: "POLICY: MCP rug-pull rule matched docker/pip lines in skills that declare no MCP server; its pin check is single-line, and the current tag is the supported track for these images"
+
+  # PE2/TM2 — sudo. Driver install, perf_event_paranoid, and cgroup limits cannot be
+  # done without it. The skills do not run it silently: SKILL.md carries a sudo
+  # protocol that makes the agent present the command and wait for the user, and
+  # tools/validate_skills.py already fails a skill whose text tells the agent to
+  # elevate without asking.
+  - id: "PE2"
+    reason: "POLICY: privileged setup is the subject of these skills; a sudo protocol gates it and validate_skills.py enforces the protocol"
```
xaleryb: The two clauses about enforcement do not hold, and this is the entry with the widest reach, so the reason is the whole audit trail for it. `DANGEROUS_CONTENT` in `validate_skills.py` is five shapes -- curl-pipe-shell, destructive-root-delete, prompt-injection, secret-exfiltration, disable-security-controls -- and its only `sudo` is inside the curl-pipe-shell regex, so nothing there enforces a protocol. The protocol as written text appears in one skill (`linux-perf/SKILL.md:39`), while `xpu-system-setup` -- the skill that actually installs drivers -- documents `--auto` as "install all without prompts (requires sudo)" at `SKILL.md:82`.

Keeping the suppression, restating what is actually true, and naming the residual risk:

```suggestion
  # PE2/TM2 - sudo. Driver install, perf_event_paranoid, and cgroup limits cannot be
  # done without it, and the elevation is the subject of these skills rather than a side
  # effect of them. Nothing enforces that mechanically: validate_skills.py checks five
  # content shapes and none of them is a sudo protocol, and the protocol itself is
  # written out in one skill (linux-perf) while the skill that actually installs drivers
  # offers `--auto` - "install all without prompts". A new skill that elevated silently
  # would therefore land accepted; worth revisiting if PE2 stops being confined to setup.
  - id: "PE2"
    reason: "POLICY: elevation is the subject of these skills - driver install, perf_event_paranoid, cgroup limits - and each documents that it needs sudo"
```

xaleryb (commented):
Ran the scanner at the pinned commit over the whole catalog, and re-ran it with variant baselines, so the numbers below are measured rather than derived from the scoring formula.

The gate reproduces: 33 skills, 17 active findings, max 35 against the 50 import limit, exit 0. Two things the run showed that are worth acting on.

**The PE2 reason is not true, and it is the only thing holding that suppression up.** `validate_skills.py` checks five content shapes and none is a sudo protocol -- its only `sudo` is inside the curl-pipe-shell regex. The protocol as text exists in `linux-perf/SKILL.md:39`; `xpu-system-setup/SKILL.md:82` documents `--auto` as "install all without prompts (requires sudo)". Suggestion inline. For context on the cost of the alternative: dropping the entry surfaces 36 findings and still exits 0 (`sglang-xpu-run` 35 -> 47, `xpu-system-setup` 15 -> 27), but its two most expensive matches are the phrase "the sglang image runs as non-root user", so blanket activation buys mostly noise. Also worth knowing before narrowing anything: `path: "SKILL.md"` puts `xpu-system-setup` at **31**, higher than fully active, because the diminishing-returns slots then fill with executable-file findings carrying the x1.3 multiplier.

**The gate cannot see severity, and the scanner already offers the field.** Six skills pass today with active HIGH findings, and `xpu-profile-unitrace` prints `LOW` in the table while carrying an unsuppressed HIGH (PE5). `risk_assessment.max_issue_severity` is in this version's JSON and is active-only -- measured: `xpu-system-setup` with four suppressed HIGH reports `MEDIUM`, `linux-perf` reports `NONE`. Reading it into `Result` would fix the table column even if you do not want it to block. Do not reach for `recommendation` instead: all 33 skills report `CAUTION`, because the reference resolver counts prose filenames (`data.npy`, `version: "1.0"`) as unresolved and fails the recommendation closed from SAFE. Same reason the comment at `tools/scan_skills.py:205-208` describes more than the code checks.

Smaller, none blocking:

- AST4 is "subprocess module call" and does not distinguish `shell=True`, so its reason asserts a property the rule cannot see -- and all 11 matches are in imported bodies that move with a pin (`linux-perf/tools/branchprob.py`, `model-config-recommend/scripts/*.py`). There are zero `shell=True` / `os.system(` under `skills/` and `tools/` today, so a one-line guard in `validate_skills.py` would make that reason enforceable rather than asserted.
- The SARIF second pass runs `check=False` and `merge_sarif` skips an unreadable per-skill file with a bare `continue`, so code scanning can under-report on a green run. Counting merged files against expected would surface it. (I checked the adjacent worry and it is fine: `and result.ok` only skips the pass for skills whose scan errored, and the merged SARIF is written at `:545` before the `return 1` at `:620`, so a threshold failure does upload its findings.)
- `CONTRIBUTING.md:230` ships the literal `@<commit>` -- the real one is at `:249`.

RP1 and LP3 I checked and would leave blanket: no skill in the catalog declares an MCP server, and the frontmatter spec has no permissions field, so both rules are categorically inapplicable here -- and RP1's stated count of 36 matches is exactly what I measured.

One measurement for the two open questions in the description: the four blanket entries together suppress 91 findings, and activating any of them leaves CI green (without PE2: 53 active, max 47; without RP1+AST4+LP3: 72 active, max 48; both exit 0). So the blanket entries are load-bearing for signal-to-noise, not for a green build. And every authored skill scores 0 with 0 active findings, so the stricter 20 is currently untested by anything in the tree.

CONTRIBUTING.md
```diff
@@ -216,7 +216,56 @@ python3 tools/sync_external.py --check           # only if you imported a skill
...
+says so. A finding the baseline accepts is not uploaded — code scanning ignores the
+suppression a SARIF file carries, so an accepted finding would arrive as an open alert.
+Its reason lives in the baseline file, which is where the audit trail belongs anyway.
+
+A finding is a question, not a verdict — the scanner does not know that `--device /dev/dri`
+is how a container sees a GPU. Read it, then either change the skill or add a rule to
+[`.skillspector-baseline.yaml`](.skillspector-baseline.yaml) saying why the pattern is right
+for this catalog. Rules, never generated fingerprints: a fingerprint is bound to the bytes
+it was made from, so it would reactivate on your next edit. Say what it is you are
+accepting — the reason is the whole value of the entry.
```
benjaminum: The security scan section is difficult to read:

Shorter and clearer suggestion

> ### The security scan
> 
> CI scans every skill with [NVIDIA SkillSpector](https://github.com/NVIDIA/SkillSpector) and
> fails a skill whose risk score is too high. (20 for skills authored in this repo, 50 for imported skills).
> 
> Run the security scan before you push:
> 
> ```bash
> # Use same skillspector commit as the CI
> SKILLSPECTOR_COMMIT=$(grep --color=never -oP 'SKILLSPECTOR_COMMIT: \K[a-f0-9]{40}' .github/workflows/skillspector.yml) 
> pip install "skillspector @ git+https://github.com/NVIDIA/skillspector.git@$SKILLSPECTOR_COMMIT"
> python3 tools/scan_skills.py your-skill-name
> ```
> Make sure to use the commit pinned in [`.github/workflows/skillspector.yml`](.github/workflows/skillspector.yml), 
> because the scanner's rule set moves between versions. The scans are static analysis only
> and do not require any API key.
> 
> To pass the security gate the risk score must be below the following thresholds:
> - Skills authored in this repo: **20 or below**
> - Imported skills: **50 or below**
> 
> A finding is a question, not a verdict — the scanner does not know that `--device /dev/dri`
> is how a container sees a GPU. Read it, then either change the skill or add a rule to
> [`.skillspector-baseline.yaml`](.skillspector-baseline.yaml) saying why the pattern is right
> for this skill. Note that rules are not pinned to the version of a skill. Explain the reason
> for each rule you add and reevaluate if you update a skill.

.skillspector-baseline.yaml
benjaminum, line 1: All rules apply to all skills, is this right?
Can we have individual rules for each skill?

benjaminum (changes requested):

.skillspector-baseline.yaml
napetrov: Taken, with your wording. I re-read the same five shapes and the only `sudo` in
`DANGEROUS_CONTENT` is inside the curl-pipe-shell regex (`validate_skills.py:207`), so
nothing there enforces a protocol; `xpu-system-setup/SKILL.md:82` documents `--auto` exactly
as you quote it. The reason now says what is true.

One thing on top of your suggestion: the residual risk you named at the end — a new skill
that elevated silently would land accepted — is now closed mechanically rather than left as
a note to revisit. PE2 carries a `skills:` list naming the six skills whose subject the
elevation already is (`linux-perf`, `performance-patterns`, `sglang-xpu-run`,
`xpu-deploy-plan`, `xpu-runtime-preflight`, `xpu-system-setup`) and TM2 names the two it
actually fires in, so a seventh skill's `sudo` arrives active instead of pre-accepted.
`tools/scan_skills.py` enforces the scope; the mechanism and the measurements are in the
reply to @benjaminum's thread on this file.

I did **not** take the `path: "SKILL.md"` narrowing, on your measurement: it puts
`xpu-system-setup` at 31, above where dropping the entry entirely leaves it, because the
diminishing-returns slots refill with x1.3 executable-file findings. Scoping by skill has
none of that effect — every per-skill score in the catalog is unchanged.

.skillspector-baseline.yaml
napetrov: No, it was not right, and it is fixed in b885cad. Thank you — this is the same hole
@xaleryb's PE2 comment reaches from the other side: an entry that names no skill also
accepts the pattern in skills nobody has written yet.

The scanner's format cannot express per-skill rules. A rule is `id`, `path`, `message`,
`reason` (`skillspector/suppression.py`), and `path` is matched against a finding's file
*within* a skill (`SKILL.md`, `scripts/foo.py`), so nothing in the schema can name the skill.
Both native routes are dead ends, measured:

- `skillspector scan skills/ --recursive --baseline .skillspector-baseline.yaml` →
  `Error: --baseline is not supported for recursive multi-skill scans; scan each sub-skill
  with its own baseline`
- one baseline per skill, inside `skills/<name>/` → `sync_external.py --check` rejects it on
  any imported skill (`FAIL … not part of the pinned upstream skill`), and imports are where
  nearly every finding is.

So the baseline carries a `skills:` list this repository defines and `tools/scan_skills.py`
enforces: before the scans start it writes each skill a temporary baseline holding only the
rules that name it, with the `skills:` key stripped — needed because SkillSpector silently
ignores fields it does not know, which is also why the file header and CONTRIBUTING now say
to run the scan through the script rather than pointing the CLI at the shared file. A scope
matching no skill fails the run rather than reading as cover:

```
FAIL bl-typo.yaml: rule 7 (TM2) is scoped to 'xpu-system-setups', which matches no skill under …/skills
```

30 of the 33 entries are scoped, which takes each skill from all 30 rules to 6–13. The
catalog result is unchanged — 17 findings active, worst score 35, every per-skill score
identical — and that is the point: it closes what the baseline promises about future skills
without moving today's answer. The effect shows up on a skill that does not exist yet, and I
measured it with one: a probe skill shipping a single `subprocess` call scores **0 with
nothing active** under the old blanket AST4 entry and **9 with the AST4 finding active** now
that AST4 names the two skills it was written for.

Three entries stay unscoped on purpose, each a claim that the rule is inapplicable to this
catalog as a whole rather than to particular skills: `RP1` (no skill declares an MCP server),
`LP3` (the frontmatter spec has no permissions field), and `PE5` (already narrow — it
suppresses four specific container flags by `message`, not the rule).

CONTRIBUTING.md
napetrov: Rewritten in 9fcdc02, close to your version. The two thresholds are a list instead of a
paragraph, the literal `@<commit>` is gone, and the section is shorter.

Two deliberate differences from the suggestion:

- The extraction uses `sed` rather than `grep -oP`: `-P` is a GNU extension, so the line
  fails on macOS's BSD grep, and this is a command a contributor pastes on whatever they
  have. Same anchoring on 40 hex characters:
  `commit=$(sed -n 's/.*SKILLSPECTOR_COMMIT: \([0-9a-f]\{40\}\).*/\1/p' .github/workflows/skillspector.yml)`
- I kept one sentence on *why* an import is held to 50 and one on `--strict-imports` for a
  pin bump. Both answer a question a contributor otherwise asks in the pull request — "why
  did my import pass at 35 when the docs say 20", and "I moved a pin, what now".

Your point about rules not being pinned to a skill version is in, as the obligation it
creates: drift-tolerant rules mean a reworded skill does not reactivate its findings, so
changing a skill means re-reading the entries scoped to it and deleting the ones its new
text no longer earns.

napetrov:
@xaleryb — every number reproduces here, including the ones I would not have guessed, and all
of it is now in b885cad and 9fcdc02 except one item I want to argue with. Measured on this
tree at the pinned commit.

**Severity is in the table now.** `Result` reads `risk_assessment.max_issue_severity` and the
table has a `worst active` column, so `xpu-profile-unitrace` reads `20 | LOW | HIGH` rather
than `LOW` alone. Confirmed active-only exactly as you describe: the field equals
max(active severity) for all 33 skills, `xpu-system-setup` with four suppressed HIGH reports
`MEDIUM`, `linux-perf` reports `NONE`. And `recommendation` is now **gone** from the table,
the JSON and the FAIL line rather than repurposed — `CAUTION` for 33/33, including skills
scoring 0 with nothing active, so it was carrying no signal at all. Not blocking on severity
for now: the score stays the gate, and the column is what stops LOW being read as safety.

**The SARIF under-report is real, and it now fails the run.** Worse than the bare `continue`
you spotted: filtering the suppressed results out before upload means a skill whose findings
were *all* accepted contributes nothing either way, so an unreadable per-skill file looked
exactly like a clean skill and a green run would have claimed coverage it did not have.
`merge_sarif` now returns the set of skills it parsed, `main` diffs that against the skills
that scanned successfully, and a gap prints `FAIL the merged SARIF is missing N scanned
skill(s): …` and exits 1, separately from the threshold failures. Verified with a
deliberately corrupt per-skill file: `count 1 parsed ['a'] incomplete ['b']`.

While in that code I found a defect of my own next door, and thanks for making me look: the
SARIF pass's `subprocess.run(..., timeout=timeout)` was unguarded, so a slow second pass
raised `TimeoutExpired` out of `scan()`, through `pool.map`, and took all 33 results with it
— the exact opposite of what the comment above it promised. Now caught, with the missing file
reported by the merge check above.

**`CONTRIBUTING.md:230`** — fixed, and the section is rewritten per @benjaminum's comment on
the same file. The commit is extracted from the workflow rather than written out at all now,
so the two cannot drift.

**AST4 and the `shell=True` guard — this is the one I want to push back on.** You are right
that the rule is "subprocess module call" and cannot see `shell=True`, and right that there
are zero `shell=True` / `os.system(` under `skills/` and `tools/` today. The guard is where I
disagree, on a measurement: `check_content_safety` calls `report.error` unconditionally
(`validate_skills.py:526-536`), unlike the mentions check right above it, which demotes to a
warning for an imported body (`:497-509`). I proved it with a synthetic imported skill —
`.source.json` present, one `shell=True` line — and `validate_skills.py` errors. Since all 11
AST4 matches are in imported bodies, as you measured, the first time that guard fires it will
be on upstream text arriving through a pin bump, where the fix cannot be made in the pull
request that is red. "Bump the pin" would become "bump the pin, then wait for upstream", and
a check that is always red teaches people to ignore red checks.

What I did instead buys the same guarantee where it matters: AST4 is now scoped to
`linux-perf` and `model-config-recommend`, so its reason claims a property of *those two*
bodies — verifiable, and I verified it — rather than of the rule, and a new skill's
`subprocess` call arrives as an active finding instead of pre-accepted (measured: 0 active
before, 9 with the finding active after). If you still want the guard, I think the right shape
is `report.warning` for an imported body and `report.error` for one written here, matching the
mentions check, and it belongs in its own pull request against `validate_skills.py` rather
than in this gate. Happy to open it.

**Left blanket, as you suggested:** `RP1` and `LP3`. Both are categorical — no skill declares
an MCP server, the frontmatter spec has no permissions field — and an entry with no `skills:`
list now *means* that, which the file header states explicitly.

On your last measurement, that every authored skill scores 0 with 0 active findings so the
stricter 20 is untested by anything in the tree: agreed, and it is why the probe-skill
measurement above is in the commit message. The threshold is untested; the scoping is what
makes sure the baseline does not quietly answer for the first skill that does test it.

napetrov:
Closing in favour of #23, which adds the same gate. One skill-security workflow, not two —
#23 is the one being carried forward.

Handing over what this branch measured, so #23 does not have to re-derive it. All numbers are
from `main` at the two pins. In priority order, what I think is worth carrying across:

1. move the pin to 2.11.2 — §1
2. read the JSON report instead of the markdown one — §3, this gates item 3
3. port the 33 reviewed suppressions — §2
4. the authored-vs-imported threshold split — §4
5. code-scanning upload of active findings only, and the `worst active` column — §5
6. the weekly cron, and the contributor-facing local-run section in `CONTRIBUTING.md` — §7

Items 1, 2 and 6 are small. Item 3 is the bulk of the work and most of it is already written.

### 1. The pin

#23 pins `939da7d4` = **SkillSpector 2.1.1 (2026-06-04)**; this branch pinned `69dcdfb7` =
**2.11.2 (2026-09-09)**. 2.1.1 has no `PE5` rule and no YARA stage at all, which changes the
answer on the finding this work was most sure about:

| skill | 2.1.1 (#23's pin) | 2.11.2 |
| --- | --- | --- |
| `torch-xpu-run` | `Score 0/100`, `Recommendation SAFE`, **0 findings** | `YR1` HIGH + `PE5` ×2 |
| whole catalog | 96 findings | 214 findings |
| `--privileged` | rule does not exist | 14 hits (`PE5`) |

`YR1` on `torch-xpu-run/SKILL.md:230` is `pip install git+https://…/transformers.git` — install
unreviewed upstream `main` into the user's environment, filed upstream as
intel/gpu-ai-skills#13. Invisible at 2.1.1.

The bump is mechanical: I ran #23's `.github/scripts/skillspector_gate.py` against markdown
reports produced by 2.11.2 and its header/location/message regexes parse them correctly.

### 2. The allowlist starts empty in practice, and the gate is loud

#23's `skillspector-allow.yml` has one entry — `xpu-system-setup` / `P2` / `SKILL.md` / match
`Hidden Instructions`. It suppresses nothing: `P2` does not appear in that skill's report at
either pin. It is inherited from amd/skills, where the external-ref provenance banner still
existed; this repository vendors imports with `.source.json` and generates no banner.

Meanwhile the gate flags **10 of 33 skills** on current `main` (measured, at #23's pin, with
that allowlist applied):

```
linux-perf              P2, P6 x2, TM2
model-can-it-fit        E2 x2
model-config-recommend  E2 x6
performance-patterns    MP3 x2
sglang-xpu-bench        SC2
sglang-xpu-run          SC2 x2
vllm-xpu-bench          SC2
xpu-model-type-detect   E2, TM1
xpu-port                AST8 (CRITICAL), AST1
xpu-system-setup        TM2 x2, TM1
```

Every one of those is a category this branch triaged with a written reason: the documented
`sudo` protocol (`TM2`), a curl against the loopback endpoint the skill just started (`SC2`),
"swap instructions" as a SIMD mnemonic (`MP3`), and `xpu-port`'s `exec(compile(…))`, which is
that tool's documented purpose and already carries a Bandit justification at the call site
(`AST1`/`AST8`, tracked rather than accepted). `.skillspector-baseline.yaml` on this branch is
the reviewed version of that list — 33 entries, each with a reason and each scoped to the
skills it applies to — and porting it into `skillspector-allow.yml` is cheaper than triaging
from zero. Sizing the port: at 2.11.2 the catalog has **39 distinct HIGH/CRITICAL
`(skill, rule, file)` tuples**, which is what the allowlist has to name once the gate is
per-finding on HIGH/CRITICAL. Two mechanical mismatches: `Suppression` requires `file`, and
several entries here are message-only (they hold across a skill's files), so those either
expand per-file or `file` becomes optional; and see the next section for the one that is not
mechanical.

### 3. The gate reads the markdown report, and that format cannot express the suppressions

Worth knowing before porting the list above, because it is a correctness limit rather than a
style question. `skillspector_gate.py` parses the markdown report, whose per-finding block is

```
### 🔴 HIGH: PE5

**Location:** `SKILL.md:65`
**Confidence:** 80%

**Message:** Privileged Container / Container Escape
```

`**Message:**` is the **pattern name**, not the matched text, and there is no snippet field. So
the finest key available to `Suppression` is `(skill, rule, file, substring-of-pattern-name)`.
Every `PE5` finding in the catalog carries that identical message.

This branch accepts four `PE5` shapes by matched text — `--device /dev/*`, `--ipc=host`,
`--net=host`, `--network host` — and deliberately leaves `--privileged` and `nsenter` active.
Two files hold both kinds:

```
sglang-xpu-run     SKILL.md   PE5  accepted 6   still-reported 6
xpu-container-run  SKILL.md   PE5  accepted 14  still-reported 4
```

Same skill, same rule, same file, same message string. An allowlist entry that accepts
render-node passthrough in those files also silences `--privileged` there — the finding this
gate most wants to keep in view.

`--format json` (what `tools/scan_skills.py` reads) carries `finding` — the matched text — plus
`location.start_line`, `risk_assessment.score` and `max_issue_severity`, so the same allowlist
becomes expressible and the regexes over prose go away. Recommend switching the gate's input
before porting the entries, otherwise the port has to be per-line or it silently over-suppresses.

### 4. Two thresholds by origin

The one design point I would not want to lose. A skill written here can be fixed in the pull
request that reports the problem, so it is held to 20 (SkillSpector's LOW band, where all ten
already sit). An imported body cannot — `sync_external.py --check` requires it to stay
byte-for-byte the pinned upstream commit — so its repair lands upstream and arrives through a
moved pin, and it is held to 50 (the scanner's own `DO_NOT_INSTALL` boundary). Same asymmetry
that already makes the mentions check and the link check warn rather than fail in an import.
`.source.json` is the origin marker. Without the split, an import's finding is either
unfixable-and-blocking or ignored; #23 currently resolves that by being advisory for
everything.

Related: when a PR *moves* a pin, that is the moment to re-read that import strictly. This
branch did it as a reported-not-blocking step (`--strict-imports`).

### 5. Platform facts worth not rediscovering

- **Code scanning ignores SARIF `suppressions`.** Uploaded 214 results, 197 carrying
  `suppressions[].justification`; all 214 came back **open**, 0 dismissed — same with
  `kind: "external"` and `kind: "inSource"`. Filter accepted findings out before upload.
- A fork's token cannot write `security-events`, so SARIF upload is branch-only; annotations
  and the step summary are what a fork PR gets. GitHub renders 10 annotations per level per
  step.
- `risk_assessment.recommendation` is **`CAUTION` for all 33 skills**, including ones scoring
  0 with nothing active, because the reference resolver counts filenames mentioned in prose as
  unresolved. Do not gate or report on it. `risk_assessment.max_issue_severity` is active-only
  and is the real signal — six skills pass today carrying an active HIGH, and
  `xpu-profile-unitrace` sits in the LOW band while carrying an unsuppressed one.
- `skillspector scan skills/ --recursive --baseline …` →
  `Error: --baseline is not supported for recursive multi-skill scans`. And a per-skill
  baseline placed inside `skills/<name>/` fails `sync_external.py --check` on any imported
  skill (`not part of the pinned upstream skill`), which is where nearly every finding is.

### 7. Two things #23 has no equivalent of

- **A scheduled full scan.** #23 triggers on `paths:` and then narrows to the skills the diff
  touched, so nothing rescans the catalog when the scanner's rules or OSV's data move — and
  dependabot will not bump this pin either, since it watches `github-actions` and the repo has
  no Python manifest. A weekly cron over all 33 is what makes a stale pin visible.
- **A contributor path.** `CONTRIBUTING.md` here gains the local invocation (commit extracted
  from the workflow so the two cannot drift), what the two thresholds mean, and what to do with
  a finding — accept it with a reason and the skills it applies to, or fix it. Without that, the
  first person to see a `::warning::` has nowhere to read.

### 8. Small ones, on #23 specifically

- The four copied files say "See `LICENSE-THIRD-PARTY.md` in the root directory"; that file
  does not exist — the licence text landed in `THIRD-PARTY-PROGRAMS.txt`.
- No `concurrency:`, no `timeout-minutes`, and no `persist-credentials: false` on either
  checkout. All four existing workflows have all three.
- It introduces a second `actions/checkout` pin (`11d5960a`, v4.4.0) alongside the repo's
  `3d3c42e5` (v7.0.1, 8 uses), and `astral-sh/setup-uv`, which is not on `main` today.
- `actionlint` and `zizmor` at this repo's pinned settings are **clean** on that branch — I
  checked, so that is not a concern.

Thanks @xaleryb and @benjaminum — the review here produced most of the measurements above, and
they outlive the branch. `enable-skillspector` stays at 9fcdc02 if any of it is wanted as a
patch against #23.

intel/gpu-ai-skills#13 is independent of this PR and stays open.

napetrov:
Closing in favour of #23, which adds the same gate. One skill-security workflow, not two —
#23's discover → matrix → aggregate shape is the one being carried forward.

The part of this branch worth keeping is already ported: **#24** is #23's commit plus one more,
so it can land either way round. What follows is the measurement record behind it, so nobody has
to re-derive it. All numbers are from `main` at the two pins. In priority order:

1. move the pin to 2.11.2 — §1, ported
2. read the JSON report instead of the markdown one — §3, ported; it gates item 3
3. the reviewed suppressions, in SkillSpector's own `--baseline` format — §2, ported (30 rules)
4. the authored-vs-imported threshold split — §4, ported
5. code-scanning upload of active findings only — §5, ported
6. the weekly cron, and the contributor-facing local-run section in `CONTRIBUTING.md` — §7, ported

What is *not* ported, deliberately: this branch's 772-line runner, its process pool and its
per-skill isolation. One skill per matrix job gives all three for free.

### 1. The pin

#23 pins `939da7d4` = **SkillSpector 2.1.1 (2026-06-04)**; this branch pinned `69dcdfb7` =
**2.11.2 (2026-09-09)**. 2.1.1 has no `PE5` rule and no YARA stage at all, which changes the
answer on the finding this work was most sure about:

| skill | 2.1.1 (#23's pin) | 2.11.2 |
| --- | --- | --- |
| `torch-xpu-run` | `Score 0/100`, `Recommendation SAFE`, **0 findings** | `YR1` HIGH + `PE5` ×2 |
| whole catalog | 96 findings | 214 findings |
| `--privileged` | rule does not exist | 14 hits (`PE5`) |

`YR1` on `torch-xpu-run/SKILL.md:230` is `pip install git+https://…/transformers.git` — install
unreviewed upstream `main` into the user's environment, filed upstream as
intel/gpu-ai-skills#13. Invisible at 2.1.1.

The bump is mechanical: I ran #23's `.github/scripts/skillspector_gate.py` against markdown
reports produced by 2.11.2 and its header/location/message regexes parse them correctly.

### 2. The allowlist starts empty in practice, and the gate is loud

#23's `skillspector-allow.yml` has one entry — `xpu-system-setup` / `P2` / `SKILL.md` / match
`Hidden Instructions`. It suppresses nothing: `P2` does not appear in that skill's report at
either pin. It is inherited from amd/skills, where the external-ref provenance banner still
existed; this repository vendors imports with `.source.json` and generates no banner.

Meanwhile the gate flags **10 of 33 skills** on current `main` (measured, at #23's pin, with
that allowlist applied):

```
linux-perf              P2, P6 x2, TM2
model-can-it-fit        E2 x2
model-config-recommend  E2 x6
performance-patterns    MP3 x2
sglang-xpu-bench        SC2
sglang-xpu-run          SC2 x2
vllm-xpu-bench          SC2
xpu-model-type-detect   E2, TM1
xpu-port                AST8 (CRITICAL), AST1
xpu-system-setup        TM2 x2, TM1
```

Every one of those is a category this branch triaged with a written reason: the documented
`sudo` protocol (`TM2`), a curl against the loopback endpoint the skill just started (`SC2`),
"swap instructions" as a SIMD mnemonic (`MP3`), and `xpu-port`'s `exec(compile(…))`, which is
that tool's documented purpose and already carries a Bandit justification at the call site
(`AST1`/`AST8`, tracked rather than accepted). `.skillspector-baseline.yaml` is the reviewed
version of that list, each entry with a reason and the skills it applies to; #24 carries it in
SkillSpector's own `--baseline` format instead of a second allowlist, which deletes the
`Suppression` class and the three regexes over markdown prose along with it. The scanner applies
the file itself and recomputes the score over what is left.

### 3. The gate reads the markdown report, and that format cannot express the suppressions

Worth knowing before porting the list above, because it is a correctness limit rather than a
style question. `skillspector_gate.py` parses the markdown report, whose per-finding block is

```
### 🔴 HIGH: PE5

**Location:** `SKILL.md:65`
**Confidence:** 80%

**Message:** Privileged Container / Container Escape
```

`**Message:**` is the **pattern name**, not the matched text, and there is no snippet field. So
the finest key available to `Suppression` is `(skill, rule, file, substring-of-pattern-name)`.
Every `PE5` finding in the catalog carries that identical message.

This branch accepts four `PE5` shapes by matched text — `--device /dev/*`, `--ipc=host`,
`--net=host`, `--network host` — and deliberately leaves `--privileged` and `nsenter` active.
Two files hold both kinds:

```
sglang-xpu-run     SKILL.md   PE5  accepted 6   still-reported 6
xpu-container-run  SKILL.md   PE5  accepted 14  still-reported 4
```

Same skill, same rule, same file, same message string. An allowlist entry that accepts
render-node passthrough in those files also silences `--privileged` there — the finding this
gate most wants to keep in view.

`--format json` (what `tools/scan_skills.py` reads) carries `finding` — the matched text — plus
`location.start_line`, `risk_assessment.score` and `max_issue_severity`, so the distinction is
expressible there. That is why #24 switches the gate's input as well as porting the entries: on
the markdown report the port would have to be per-line or it silently over-suppresses.

### 4. Two thresholds by origin

The one design point I would not want to lose. A skill written here can be fixed in the pull
request that reports the problem, so it is held to 20 (SkillSpector's LOW band, where all ten
already sit). An imported body cannot — `sync_external.py --check` requires it to stay
byte-for-byte the pinned upstream commit — so its repair lands upstream and arrives through a
moved pin, and it is held to 50 (the scanner's own `DO_NOT_INSTALL` boundary). Same asymmetry
that already makes the mentions check and the link check warn rather than fail in an import.
`.source.json` is the origin marker. Without the split, an import's finding is either
unfixable-and-blocking or ignored; #23 currently resolves that by being advisory for
everything.

Related: when a PR *moves* a pin, that is the moment to re-read that import strictly. This
branch did it as a reported-not-blocking step (`--strict-imports`).

### 5. Platform facts worth not rediscovering

- **Code scanning ignores SARIF `suppressions`.** Uploaded 214 results, 197 carrying
  `suppressions[].justification`; all 214 came back **open**, 0 dismissed — same with
  `kind: "external"` and `kind: "inSource"`. Filter accepted findings out before upload.
- A fork's token cannot write `security-events`, so SARIF upload is branch-only; annotations
  and the step summary are what a fork PR gets. GitHub renders 10 annotations per level per
  step.
- `risk_assessment.recommendation` is **`CAUTION` for all 33 skills**, including ones scoring
  0 with nothing active, because the reference resolver counts filenames mentioned in prose as
  unresolved. Do not gate or report on it. `risk_assessment.max_issue_severity` is active-only
  and is the real signal — six skills pass today carrying an active HIGH, and
  `xpu-profile-unitrace` sits in the LOW band while carrying an unsuppressed one.
- `skillspector scan skills/ --recursive --baseline …` →
  `Error: --baseline is not supported for recursive multi-skill scans`. And a per-skill
  baseline placed inside `skills/<name>/` fails `sync_external.py --check` on any imported
  skill (`not part of the pinned upstream skill`), which is where nearly every finding is.

### 7. Two things #23 has no equivalent of

- **A scheduled full scan.** #23 triggers on `paths:` and then narrows to the skills the diff
  touched, so nothing rescans the catalog when the scanner's rules or OSV's data move — and
  dependabot will not bump this pin either, since it watches `github-actions` and the repo has
  no Python manifest. A weekly cron over all 33 is what makes a stale pin visible.
- **A contributor path.** `CONTRIBUTING.md` here gains the local invocation (commit extracted
  from the workflow so the two cannot drift), what the two thresholds mean, and what to do with
  a finding — accept it with a reason and the skills it applies to, or fix it. Without that, the
  first person to see a `::warning::` has nowhere to read.

### 8. Small ones, on #23 specifically

- The four copied files say "See `LICENSE-THIRD-PARTY.md` in the root directory"; that file
  does not exist — the licence text landed in `THIRD-PARTY-PROGRAMS.txt`.
- No `concurrency:`, no `timeout-minutes`, and no `persist-credentials: false` on either
  checkout. All four existing workflows have all three.
- It introduces a second `actions/checkout` pin (`11d5960a`, v4.4.0) alongside the repo's
  `3d3c42e5` (v7.0.1, 8 uses), and `astral-sh/setup-uv`, which is not on `main` today.
- `actionlint` and `zizmor` at this repo's pinned settings are **clean** on that branch — I
  checked, so that is not a concern.

The first three items in §8 are fixed in #24; `actionlint` and `zizmor` are clean there at this
repository's pinned digests, and the scan passes 33/33 on the current catalog.

Thanks @xaleryb and @benjaminum — the review here produced most of the measurements above, and
they outlive the branch. `enable-skillspector` stays at 9fcdc02 if anything not in #24 is wanted
later.

intel/gpu-ai-skills#13 is independent of this PR and stays open.

PR #23, opened 2026-09-14
Add a CI workflow for static validation on skills.

mzweilin:
## What this changes

This pull request introduces SkillSpector static analysis tooling and infrastructure to the repository, enabling automated security scanning of skills and providing a mechanism to suppress known false positives. The main changes include adding scripts for selecting and gating skills for scanning, a YAML allowlist for suppressing benign findings, and updating CI runner labels. The copyright notice and the original license is included.

We have configured the workflow to be advisory-only for PR reviewers. It won't stop a PR from being merged.

## Checklist

- [ ] I checked `description` against requests a user would really type — see
      CONTRIBUTING.md.
- [ ] `python3 tools/validate_skills.py` passes locally.


.github/workflows/skillspector.yml
```diff
@@ -0,0 +1,204 @@
...
+# This workflow is ADVISORY ONLY: SkillSpector findings never fail CI. Each
+# changed skill with un-allowlisted HIGH/CRITICAL findings (or a tool error)
+# raises a ::warning:: annotation plus a job-summary entry so reviewers are
+# alerted that action is needed, without blocking the merge. Keep this check
+# non-required in branch protection to preserve that behavior.
+
+on:
+  push:
+    # Test it on the dev branch before we merge to main
+    branches: [main, mzweilin/ci_skill_spector]
```
mzweilin, line 30: Only keep `branches: [main]` before merging. 
`mzweilin/ci_skill_spector` is for dev only.

xaleryb:
The shape of this is right, and several choices here are better than what #19 had: one scan per skill with `fail-fast: false` and a per-skill artifact, so a suppression is scoped by construction rather than by a repo-invented YAML key; `contents: read` only; no conditional branches for forks, so a fork PR gets the same answer as a branch; the scan narrows to the skills a diff touched and re-scans everything when the machinery changes (`changed_skills.py:119-121`), and it fails safe — an unresolvable base or a failed `git diff` scans all skills (`:103-108`, `:111-117`). Copied code carries its MIT header and the text landed in `THIRD-PARTY-PROGRAMS.txt`. Advisory-first is the right call for a new scanner.

Five small things, none of them a reason to hold this:

**1. The dev branch is still in the push trigger.** `branches: [main, mzweilin/ci_skill_spector]` (`.github/workflows/skillspector.yml:30`) — the comment above it names the condition for removing it.

**2. The licence header points at a file that doesn't exist.** Four files say "See `LICENSE-THIRD-PARTY.md` in the root directory" (`skillspector.yml:3`, `skillspector-allow.yml:3`, `scripts/changed_skills.py:5`, `scripts/skillspector_gate.py:5`); the text is in `THIRD-PARTY-PROGRAMS.txt`, which this PR adds. One `sed`.

**3. Three settings the other five workflows all have.** No `concurrency:` (5 of 5 on `main`), no `timeout-minutes:` on any of the three jobs (7 of the 8 jobs on `main` set one), no `persist-credentials: false` on either checkout (all 8 checkouts on `main` set it — `:56-60`, `:94-95`). `concurrency` isn't only consistency: the matrix fans out to one job per changed skill, so a series of pushes queues whole fan-outs. The idiom is `validate.yml:22-24`.

**4. The one allowlist entry describes a different repository.** It came across with the workflow, and three of its four factual claims don't hold here (`skillspector-allow.yml:23-41`):

- "the provenance banner … (lines 18-23)" with a "`path` line" whose keyword hit is `system` — the note here is a single line and has no `path` line: `Modified by intel/skills: upstream repository-relative paths rewritten to resolve where this skill installs. Provenance: .source.json` (`tools/validate_skills.py:184-186`, and in the file at `skills/xpu-system-setup/SKILL.md:5`). None of the pattern's keywords (`system|instructions?|ignore|POST|GET|send|transmit`) appear in it, so the mechanism the entry describes isn't available here.
- "All 23 external-ref pointers carry the byte-identical banner" — 23 skills carry `.source.json`, but the note is in 7 files across 5 of them: exactly the files their manifests list under `modified-files`, because `tools/sync_external.py` writes it only where it rewrote a path.
- "it is omitted entirely once pointers are vendored with `.source.json`" — the reverse: the note exists *because* they are vendored.
- The one claim that does hold, under another name: the note is mandatory, enforced by `check_imports` (`tools/validate_skills.py:1052-1057`), which errors with "records this file as modified, but the file itself does not say so" — not by a `check_external_refs`, which doesn't exist here.

I'd delete the entry and let the allowlist start empty rather than reword it. It's the only content claim in the PR and every later entry will be written from it as a template — and whether `P2` fires on this skill at this pin isn't something either of us has measured.

**5. Two paths where the gate reports the wrong thing.** `exit 2` is handled well — it says "SkillSpector errored" and skips the gate (`:135-142`), so that's clear. The other two aren't:

- A report that exists but is empty or truncated parses to zero findings, the gate returns 0 (`scripts/skillspector_gate.py:174`, `:206-210`) and the summary prints `:white_check_mark: no un-allowlisted HIGH/CRITICAL findings` (`:161`). A `[ -s "$report" ]` check before the gate separates "scanned and clean" from "nothing to read".
- Any other non-zero exit that leaves no report falls through to the gate, which returns 1 via `Report not found` (`:170-172`), and the workflow announces it as `::warning title=SkillSpector findings::… un-allowlisted HIGH/CRITICAL` (`:155`). The stderr line is in the log, but the annotation says the opposite of what happened. The aggregate job already gets this right for a crashed job (`:199-203`).

One more: keep this check non-required in branch protection, as `:21-25` asks — worth stating in the PR description so it isn't lost.

On carrying over from #19, the order matters. The pin here (`939da7d4`, `:125`) isn't the one that PR's triage was measured against, and its driver read `--format json` (`tools/scan_skills.py:244`), where a suppression key gets the matched text and a line number; this gate keys on the rule, the file, and a substring of `**Message:**` (`skillspector_gate.py:54-58`, `:77-84`). If `**Message:**` carries the pattern name rather than the matched text — that PR's closing comment reports measuring exactly that, and I haven't re-run it — then several findings in one file are indistinguishable to the allowlist and porting entries before switching format over-suppresses. So: move the pin, confirm what the message field contains, then port entries. Two numbers in that closing comment are off, so they don't get inherited: its baseline has **30** entries, 24 of them scoped, not 33 (33 is the skill count); and this repo does write a note into vendored files, so "generates no banner" isn't a safe premise. Its finding counts and claims about which rules exist at either pin were measured on that branch — treat them as its author's measurements, not as facts about this PR.

xaleryb:
Update — mzweilin/skills#1 is stacked on this branch and predates this comment; four of the five items above are already fixed there: the licence header in all four files, `concurrency`/`timeout-minutes`/`persist-credentials` plus the checkout pin folded back to `3d3c42e5`, the allowlist file deleted, and both reporting paths (`exit ≥ 2` → `::error` with no gate call, `execution_successful=false` fails the skill, `--expect` fails the SARIF merge). Only the dev branch in the push trigger (`skillspector.yml:30`) is left.

My ordering advice is moot in the best way: that PR deletes the markdown parsing instead of working around it, and this branch's one-skill-per-job shape is what makes that possible — the scanner refuses a shared `--baseline` only for a recursive scan. The custom half of the gate and #19's 772-line runner both go away.

Two claims in its description are wrong, in case they get quoted onward; neither changes its conclusions. "generates no banner" — it does: `REWRITE_NOTE` (`tools/validate_skills.py:184-186`) is written into the 7 files the manifests list under `modified-files`, enforced by `check_imports` (`:1052-1057`). "all four existing workflows" — there are five on `main`, with `timeout-minutes` on 7 of the 8 jobs.

On the call it leaves to maintainers — advisory to blocking on score, 20 authored / 50 imported: I'd take it. Both numbers are the scanner's own band edges, `.source.json` is already the origin marker the repo keys on, and blocking on score while only annotating an active HIGH/CRITICAL below it is the right split — a check that cannot fail gets read once. One question first: when a moved pin pushes an import past 50, the repair is upstream, `sync_external.py --check` forbids touching the body, and the pin-bump PR is the one that goes red — the argument that PR's author made against a `shell=True` guard on #19. If the answer is a TRACKED baseline entry naming the upstream fix, that belongs in `CONTRIBUTING.md` beside the thresholds.

xaleryb:
Update from the first actual run, now that the workflows are authorised: [run 34882348713](https://github.com/intel/skills/actions/runs/34882348713) — 34 jobs, all green, ~50s, on `refs/pull/23/merge`. Four facts out of the logs.

- **The findings warning never becomes an annotation.** Ten skills tripped the gate (`model-config-recommend` 11, `linux-perf` 4, `xpu-system-setup` 3, five more at 2, two at 1) and all ten emitted `::warning title=SkillSpector findings`. All 33 check runs carry exactly 2 annotations, both from actions. `cat "$report"` (`skillspector.yml:131`) prints a report whose last byte isn't a newline, so the following `echo "::warning …"` isn't at column 0 and Actions doesn't parse it — they appear glued into one line in the log. The job summary is unaffected, which is why it looks fine. `printf '\n'` after the `cat` fixes it; mzweilin/skills#1 fixes it by deleting the `cat`.
- **The allowlist suppressed nothing, on all 33 skills.** Every passing job logs `(0 allowlisted)`, and `xpu-system-setup` — the skill its only entry names — is one of the ten that failed the gate, with 3 un-allowlisted findings.
- **`branches: [main, mzweilin/ci_skill_spector]` (`:30`) has never matched.** The branch is `ci_skill_spector`; `mzweilin/` is the fork owner, not part of the ref. That fork has zero push-event runs, so this was the workflow's first execution anywhere.
- **66 warnings, two per job.** `actions/checkout@11d5960a` and `actions/upload-artifact@ea165f8d` are Node20; `setup-uv` with no inputs finds no Python manifest here, so *"the cache will never get invalidated"*. `main` uses neither action, so both pins are new.

Two more from the same run, at pin `939da7d4`: `xpu-container-run`, which documents `--privileged`, `--net=host` and `--ipc=host`, comes back `0/100`, `SAFE`, zero findings — no `PE5` rule at this version. And score band and finding severity disagree on three skills: `xpu-deploy-plan` is 52/HIGH with no HIGH/CRITICAL finding, `vllm-xpu-bench` and `sglang-xpu-bench` are 25/MEDIUM with one each.

PR #24, opened 2026-09-15
SkillSpector: v2.11.2 pin, reviewed baseline, JSON gate with origin thresholds (on top of #23)

napetrov:
**Carries #23's commit unaltered.** The first commit here is `00247b5e` *Add a CI workflow for static validation on skills.*, authored **and** committed by @mzweilin — same SHA as #23's head, not rebased and not cherry-picked. My five commits sit on top of it, so this branch is #23 plus the changes described below and the two pull requests cannot land independently.

@mzweilin offered to close #23 and take this one instead. If that is the call, please use **Rebase and merge**, not Squash: a squash would collapse `00247b5e` into a single commit authored by me, and the workflow in it is not mine. Landing #23 first works too — this then rebases onto `main` and only my commits remain.

#23 is the one to land first, so its commit reaches `main` through @mzweilin's own pull request; this branch then rebases onto `main` and `pr23-base` is deleted. Merging #24 instead would put that commit on `main` through a rebase or squash of mine, which is the wrong attribution for a workflow I did not write.

Context: intel/skills#19 was a parallel attempt at the same gate. It is being closed in favour of #23; this is the part of it worth keeping, rebuilt on this branch's shape. #23's discover → matrix → aggregate structure is kept as-is, and it makes several things in #19 unnecessary: one skill per job replaces its process pool, its per-skill isolation and most of its 772-line runner, none of which is ported.

Every number below is measured on this tree, at the pin this PR sets.

## 1. The pin moves to v2.11.2

`939da7d4` is v2.1.1 (2026-06-04), which has **no `PE5` rule and no YARA stage at all**.

| | v2.1.1 | v2.11.2 (`69dcdfb7`, tag-verified) |
| --- | --- | --- |
| whole catalog, unsuppressed | 96 findings | 214 findings |
| `torch-xpu-run` | `Score 0/100`, `SAFE`, 0 findings | `YR1` HIGH + `PE5` ×2 |
| `--privileged` | rule does not exist | 14 hits |

`YR1` on `torch-xpu-run/SKILL.md:230` is `pip install git+https://…/transformers.git` — install unreviewed upstream `main` into the user's environment. Filed upstream as intel/gpu-ai-skills#13, left unsuppressed here on purpose.

## 2. Suppressions move into SkillSpector's own baseline format

`skillspector scan --baseline` already drops the findings a rule matches, keeps the rule's `reason` with them, and recomputes the score over what is left. Using it deletes the custom half of the gate: `.github/skillspector-allow.yml`, the `Suppression` class, and the three regexes over markdown prose. `.skillspector-baseline.yaml` holds 30 rules, each with a reason and split into POLICY (inherent to what this catalog documents) and TRACKED (real, not fixed here, with the fix named).

Nothing is lost by dropping the allowlist: its single entry (`xpu-system-setup` / `P2` / `SKILL.md`, match `Hidden Instructions`) matches nothing at either pin. **Correction to an earlier version of this description**, which said the entry came from amd/skills: it did not — that repository's allowlist has 14 entries and all of them are `local-ai-use`/`serving-llms-on-instinct`. The entry is @mzweilin's, written against a real earlier-version finding on this catalog, and the reason it carries explains why it stopped matching: it targeted the provenance banner `tools/sync_external.py` generated, and that banner is "omitted entirely once pointers are vendored with `.source.json`", which is what this catalog now does. Re-measured at 2.11.2 across all 33: `P2` fires once in the whole catalog, on `linux-perf/references/flow-d.md:1` (`<!-- (C) 2026 Intel Corporation, MIT license -->`), and not in `xpu-system-setup`. The file format the allowlist uses is what came from amd/skills; I conflated the two.

**Per-skill scoping.** The rule schema is `id` / `path` / `message` / `reason`, and `path` matches a finding's file *within* a skill (`SKILL.md`, `scripts/foo.py`) — nothing in it can name a skill, so one shared file reaches all 33. `.github/scripts/skillspector_baseline.py` adds a `skills:` list the scanner ignores and emits one skill's rules per scan; a scope matching no skill on disk fails the run rather than reading as cover for a typo. Measured with a probe skill shipping one `subprocess` call: **score 0, nothing active** under a blanket `AST4` entry, **9 with the finding active** once `AST4` names the two skills it was written for.

Two native alternatives were tried and neither works: the CLI refuses a shared baseline for a `--recursive` multi-skill scan, and a per-skill baseline placed inside `skills/<name>/` fails `sync_external.py --check` on any imported skill — which is where nearly every finding is.

## 3. The gate reads the JSON report, not the markdown one

This is a correctness fix, not a style preference. The markdown per-finding block is:

```
### 🔴 HIGH: PE5

**Location:** `SKILL.md:65`
**Confidence:** 80%

**Message:** Privileged Container / Container Escape
```

`**Message:**` is the **pattern name**, and there is no matched text and no snippet anywhere in the file. So the finest key available to the old allowlist was `(skill, rule, file, substring-of-pattern-name)`, and every `PE5` finding in this catalog carries that identical message — while the four flags this repository accepts (`--device /dev/*`, `--ipc=host`, `--net=host`, `--network host`) and the two it will not (`--privileged`, `nsenter`) share a rule, a file, and often a paragraph:

```
sglang-xpu-run     SKILL.md   PE5  accepted 6   active 6
xpu-container-run  SKILL.md   PE5  accepted 14  active 4
```

An allowlist entry accepting render-node passthrough in those files would also silence `--privileged` there. The JSON carries `finding` (the matched text) and `location.start_line`, so the distinction is expressible — and the gate gets shorter, because the regexes go away.

Two more fields worth naming: the gate reports `risk_assessment.max_issue_severity`, which is computed over the active findings only, and deliberately never reads `recommendation` — it is `CAUTION` for **all 33 skills**, including ones scoring 0 with nothing active, because the reference resolver counts filenames mentioned in prose as unresolved.

## 4. Two thresholds by origin, and what blocks

A skill written here can be fixed in the pull request that reports the problem. An imported skill is upstream's text, kept byte-for-byte because `sync_external.py --check` compares it to the pinned commit, so its repair lands upstream and arrives through a moved pin — the same asymmetry that already makes the mentions check and the link check warn rather than fail in an imported body.

| origin | fails above | why that number |
| --- | --- | --- |
| written here (10 skills) | **20** | SkillSpector's LOW/`SAFE` band — where all ten already sit |
| imported (23 skills) | **50** | the boundary above which the scanner itself says `DO_NOT_INSTALL` |

Neither number was picked here; both are the scanner's own band edges. `.source.json` is the origin marker.

**This is the one policy change from #23's advisory-only stance, and it is the maintainers' call.** What blocks: a score above the origin's threshold, or a scan that did not complete. What does not block: an active HIGH/CRITICAL finding below the threshold — it is annotated on the diff and written to the job summary, because blocking on it would make "bump an import's pin" mean "bump the pin and then wait for upstream", and a check that is always red teaches people to ignore red checks. If you would rather keep the whole thing advisory, `continue-on-error: true` on the gate step is the switch, and the reporting is unchanged.

Today's catalog would not be blocked either way: gate passes **33/33**.

## 5. Reporting, and refusing to claim coverage

- **Code scanning**: one merged SARIF, repository-root paths, `category: skillspector` so it cannot collide with CodeQL's alerts. Non-fork and public only, the same guard `codeql.yml` carries — a fork's token cannot write `security-events`, so annotations and the job summary are what a fork PR gets.
- **Suppressed findings are deliberately not uploaded.** SkillSpector writes each one out with the baseline `reason` in `suppressions[].justification`, which reads like it should arrive as a dismissed alert. Measured against the API, it does not: **214 results uploaded, 197 carrying a justification, all 214 came back open**, with `kind: external` and `kind: inSource` alike. Uploading them would bury the 17 that want reading under 197 nobody will act on.
- `--expect` on the merge is there because dropping the suppressed findings means a skill whose findings were *all* accepted contributes nothing to the merged file — so a missing or unreadable per-skill SARIF looks exactly like a clean skill. A gap fails the job. Verified by deleting one file and by corrupting another.
- **A weekly cron**, because the `paths:` filter and the changed-skills narrowing together mean nothing re-reads a skill when the *scanner's* rules or its OSV data move, and dependabot will not bump this pin (it watches `github-actions`, and this repository has no Python manifest).

## 6. Conventions and small fixes

- `concurrency:`, `timeout-minutes:` and `persist-credentials: false` — all four existing workflows have all three, this one had none.
- `actions/checkout` on `3d3c42e5` (v7.0.1), the pin the other eight uses in this repository share, instead of introducing a second one.
- The third-party notice now points at `THIRD-PARTY-PROGRAMS.txt`, which exists; `LICENSE-THIRD-PARTY.md` does not.
- `changed_skills.py`: `INFRA_PATHS` tracks the new scripts and the baseline, and skill discovery requires a `SKILL.md` so a stray directory cannot fan out a job that can only fail.
- `CONTRIBUTING.md` documents the local run, with the pin extracted from the workflow so the version a contributor runs is the version CI runs.

## Measured

| | |
| --- | --- |
| skills scanned | 33, gate passes 33/33 |
| with the baseline | 26 at score 0, worst **35** against a limit of 50, **17** active findings |
| unsuppressed, same pin | 214 findings, 12 skills at `DO_NOT_INSTALL` (7 CRITICAL, worst 100/100) |
| every non-zero score | an imported skill; all ten authored here are at 0 |
| what stays active | `PE5` ×14 (`--privileged` 13, `nsenter` 1, in a sentence telling you not to use it), `RA2` ×2 (`systemctl enable --now docker`), `YR1` ×1 |

Also verified: `actionlint` clean and `zizmor --min-severity=medium --min-confidence=medium` clean at this repository's pinned digests; all four scripts run under `uv run` with their inline dependency headers; `validate_skills.py`, `run_evals.py --validate` and `sync_external.py --check` unaffected and passing; the full-rescan trigger fires for the new infra paths; the negative paths exercised (over-threshold score, `execution_successful=false`, unknown skill, bad `skills:` scope, non-list scope, missing and corrupt per-skill SARIF).

Happy to split any of the six sections into its own commit, or to drop §5 entirely if you would rather land the pin, the baseline and the JSON gate first — those three are the ones that change what the scan can see.

## Since review

- `branches: [main]`, comment dropped (@mzweilin) — 18b84b5.
- The allowlist provenance claim in §2 above was wrong and is corrected there (@mzweilin).
- `RP1` scoped to the ten skills its 36 findings were measured in; `--net=host` and `--network host` scoped to the one skill each (5 and 1 findings). `LP3` and the other two `PE5` flag globs stay unscoped, and the file argues each (@xaleryb) — 18b84b5, a17a67c. 27 of 30 rules now name their skills. Measured state unchanged: 33/33, 17 active, 197 suppressed, worst 35.
- The gate emits one `::error` carrying the headline whenever it fails, so a skill crossing its limit on MEDIUM findings alone lands on the diff and not only in the step summary; it takes one of the level's ten annotation slots and the dropped-count notice accounts for it (@xaleryb) — 18b84b5.
- `CONTRIBUTING.md`'s security-scan section is 15 lines shorter: thresholds as a list, an unscoped rule described as applying to every skill, the fingerprint aside gone (@benjaminum) — 25bd85d.

Re-validated on a merge of this branch with current `main` (which has since gained #22): merge is clean, `validate_skills.py` / `run_evals.py --validate` / `sync_external.py --check` / `lint_task_leakage.py --self-test` / `lint_task_leakage.py --fail-on-leak 5` all pass, gate 33/33 with 15 warning annotations and no error, `actionlint` and `zizmor` clean.


.github/workflows/skillspector.yml
```diff
@@ -0,0 +1,336 @@
...
+# What fails: a skill scoring above the threshold its origin is held to (20 for
+# a skill authored here, 50 for an imported body -- see skillspector_gate.py for
+# why the two differ), or a scan that did not complete. An active HIGH/CRITICAL
+# finding below the threshold is reported as a ::warning:: annotation and in the
+# job summary, because blocking on it would make "bump an import's pin" mean
+# "bump the pin and then wait for upstream".
+
+on:
+  push:
+    # Test it on the dev branch before we merge to main
```
mzweilin:
```suggestion
```

.github/workflows/skillspector.yml
```diff
@@ -0,0 +1,336 @@
...
+    branches: [main, mzweilin/ci_skill_spector]
```
mzweilin:
```suggestion
    branches: [main]
```

mzweilin (changes requested):
I like your changes. Just want to clarify that the allow list is not copied from AMD. It is a false positive on the `intel/xpu-system-setup` skill generated by an earlier version of SkillSpector.

You may merge #24 and close #23 for simpicity.

xaleryb (approved):
Approving. I re-measured this from [run 34929676992](https://github.com/intel/skills/actions/runs/34929676992) rather than reading it off the diff, and every number in the body held: 33 skills, gate 33/33, 26 at score 0, worst **35** against the imported limit of 50 (`xpu-container-run`, `sglang-xpu-run`), 17 active and 197 suppressed, all ten skills authored here at 0 — every non-zero score is an import.

The part only a run could settle is the annotation path, and it works here. On #23 all 33 check runs carried exactly two annotations, both from actions, because `cat "$report"` left stdout mid-line and the `::warning` after it was never parsed. Here the counts differ per skill and carry a location and the matched text:

```
skills/sglang-xpu-run/SKILL.md:213   SkillSpector PE5 (HIGH)   sglang-xpu-run: --privileged
```

Dropping the markdown report removes that failure mode by construction, and the ten-per-level cap is handled explicitly rather than discovered later. Node20 warnings go from 66 per run to 33 — `upload-artifact@ea165f8d` is the only one left now that checkout is on v7.0.1.

Two notes, neither blocking:

- `RP1`'s baseline entry has no `skills:` list, but its reason states the condition it relies on — *"matched docker/pip lines in skills that declare no MCP server"*. A skill that does declare one would have a genuine rug-pull finding silenced by the same entry. Scoping it to the skills it was measured on keeps the reason and the rule saying the same thing.
- `skillspector_gate.py` prints `FAIL` to stderr, and annotations are emitted for HIGH/CRITICAL only. So a skill authored here that crosses 20 on MEDIUM findings alone produces a red check with no annotation on the diff — the reason is in the step summary only. An `::error` carrying the headline would close it.

Both are follow-ups. `mzweilin`'s two lines on the push trigger are still unapplied at `ad440e7a`, and his review is the one that has to clear — merge timing is yours and his.

CONTRIBUTING.md
```diff
@@ -218,6 +218,41 @@ python3 tools/sync_external.py --check           # only if you imported a skill
...
+```
+
+The commit is read out of the workflow so the version you run is the version CI runs.
+
+Two thresholds, because the two kinds of skill can act on a finding differently. A skill
+written here fails above a score of **20** — SkillSpector's own `SAFE` band — because a
+finding in it can be fixed in the pull request that reports it. An imported skill fails
+above **50**, the boundary above which the scanner itself says `DO_NOT_INSTALL`, because
+its body is kept byte-for-byte the pinned upstream commit and the repair has to land
+upstream. Active HIGH/CRITICAL findings below the threshold are reported, not failed.
```
benjaminum: This file is already quite wordy. Highlight the two threshold as a list with two entries here.

CONTRIBUTING.md
```diff
@@ -218,6 +218,41 @@ python3 tools/sync_external.py --check           # only if you imported a skill
...
+
+If the finding is real, fix it. If it is a false positive or a pattern this catalog
+documents on purpose, add a rule to [`.skillspector-baseline.yaml`](.skillspector-baseline.yaml)
+with a `reason` — and list your skill under `skills:`, because a rule that names no skill
+also accepts the same pattern in a skill nobody has written yet. The file is the audit
```
benjaminum: > because a rule that names no skill
also accepts the same pattern in a skill nobody has written yet"

Does this mean that rule applies to all skills? If so phrase it like that.

CONTRIBUTING.md
```diff
@@ -218,6 +218,41 @@ python3 tools/sync_external.py --check           # only if you imported a skill
...
+trail, so a suppression that cannot be justified in a sentence has nowhere to hide, and a
+rule is not pinned to a version of your text: it keeps applying after the surrounding
+paragraph is reworded, which is why the baseline uses drift-tolerant rules rather than
+generated fingerprints.
```
benjaminum: Simplify more.

The file is the audit trail, each suppression must be justified. Note that rules are not pinned to a specific version of your skill.

benjaminum (changes requested):

.github/workflows/skillspector.yml
napetrov: Both suggestions applied in 18b84b5 — `branches: [main]` and the comment with it.

You are right about the allow list and I have corrected the pull request description. I checked `amd/skills/.github/skillspector-allow.yml`: 14 entries, every one of them `local-ai-use` or `serving-llms-on-instinct`, no `xpu-system-setup` and no `P2`. The file *format* is what came from there; the entry is yours, and I conflated the two.

For the record, re-measured at 2.11.2 across all 33 skills: `P2` fires exactly once in the catalog, on `linux-perf/references/flow-d.md:1` (`<!-- (C) 2026 Intel Corporation, MIT license -->` — the licence header, suppressed by its own baseline entry), and not in `xpu-system-setup`. That matches what your own `reason` predicted: the banner it matched is "omitted entirely once pointers are vendored with `.source.json`", which is what the catalog now does. So the entry is safe to drop because the finding is gone, not because it was ever wrong.

On merging this one and closing #23 — thanks, but I would rather keep the order the other way round: land #23 as yours, and I rebase this onto `main` after it. Merge commits are disabled on this repository and a squash of #24 would collapse your commit into one authored by me, which is the wrong attribution for the workflow you wrote. `pr23-base` exists only so this pull request's diff shows my delta rather than both commits; it gets deleted once #23 lands.

napetrov:
Both notes applied in 18b84b5 rather than deferred — thanks, the first one was a real hole.

**`RP1`.** Scoped to the ten skills its findings were measured in. Unsuppressed scans of all 33 at 2.11.2 put `RP1` in exactly 10 skills / 36 findings: `llamacpp-xpu-run` 7, `xpu-container-run` 7, `sglang-xpu-run` 6, `model-config-recommend` 4, `vllm-xpu-run` 4, `vllm-xpu-bench` 2, `vllm-xpu-profile` 2, `xpu-port` 2, `torch-xpu-run` 1, `xpu-deploy-plan` 1 — those are the names now in the entry. Also confirmed the premise still holds catalog-wide: no skill declares an MCP server today (the only case-insensitive `mcp` hit in `skills/` is the string `Memcpy` in a profiler table), which is exactly why the entry should not be the thing that decides for a skill that later does.

Scoping it changes nothing measured: still 33/33, 26 at score 0, 17 active, 197 suppressed, worst 35. `skillspector_baseline.py` emits 12 rules for `sglang-xpu-run` and 5 for `dpnp-io`, with `RP1` absent from the second.

`LP3` stays unscoped, deliberately, and the file now says so where the asymmetry is visible: its reason is a property of the frontmatter spec rather than of any skill's text, so it holds for every skill including ones nobody has written yet. When the spec grows a permissions field, that arrives as a pin bump and a new rule id, not as a suppression that quietly widened.

**The stderr-only failure.** Fixed: the gate now emits one `::error file=skills/<skill>/SKILL.md,title=SkillSpector gate::` carrying the headline and the reason whenever `failures` is non-empty, so the case you describe lands on the diff. Verified with a synthetic report — score 30, two MEDIUM findings, no HIGH — against an authored skill:

```
::error file=skills/dpnp-io/SKILL.md,title=SkillSpector gate::dpnp-io: score 30/20 (authored, band MEDIUM), worst active MEDIUM, 2 active, 2 suppressed -- score 30 is above the limit of 20 for an authored skill
```

exit 1, and the same report against an imported skill (limit 50) emits no annotations at all. It takes one of the level's ten slots, so the per-finding loop and the dropped-count notice now budget for it: 12 HIGH findings on a failing skill produce 10 `::error` (1 gate + 9 findings) and a `::notice` saying 3 were not rendered.

CONTRIBUTING.md
napetrov: Done in 25bd85d — the two thresholds are a list now:

- **A skill written here fails above 20** — SkillSpector's `SAFE` band. A finding in it can be fixed in the pull request that reports it.
- **An imported skill fails above 50** — where the scanner itself says `DO_NOT_INSTALL`. Its body stays byte-for-byte the pinned upstream commit, so the repair lands upstream.

CONTRIBUTING.md
napetrov: Yes — it applies to every skill in the catalog, including ones not written yet. Phrased that way now: "list your skill under `skills:` — a rule without it applies to every skill in the catalog."

CONTRIBUTING.md
napetrov: Taken almost verbatim in 25bd85d: "The file is the audit trail, so each suppression must be justified. Rules are not pinned to a specific version of your skill: they keep applying after the surrounding text is reworded." The fingerprint aside is gone — the reasoning for it lives in the header comment of `.skillspector-baseline.yaml`, which is where someone editing the file will be. Section is 15 lines shorter.

napetrov:
Follow-through on the same principle, a17a67c: `--net=host` and `--network host` are now scoped too. Measured spread of the four accepted flags across the catalog:

```
--device /dev/    29 findings in 8 skills
--ipc=host        30 findings in 9 skills
--net=host         5 findings in 1 skill   (xpu-container-run)
--network host     1 finding  in 1 skill   (sglang-xpu-run)
```

The bottom two were blanket entries accepting the host network stack in 31 skills that never asked for it. The top two stay unscoped deliberately, and the file says why: every XPU container passes a render node through and every multi-process server needs host IPC, so scoping them would make a new container skill's first scan red for the pattern this catalog exists to document. Still 33/33, 17 active, 197 suppressed.

That leaves three unscoped entries in the file: those two `PE5` flag globs and `LP3`. Each accepts a specific matched string or a property of the spec rather than a rule id, so the blast radius of the loosest one is "the literal `--ipc=host`", not "privilege escalation".

.github/workflows/skillspector.yml
napetrov: Applied in 18b84b5 — comment and dev branch both gone, the trigger is `branches: [main]`.

benjaminum (approved):

napetrov:
Base is back on `main` and the mirror branch `pr23-base` is deleted, so #23's commit is visible here where it belongs: `00247b5e`, *Add a CI workflow for static validation on skills.*, author **and** committer @mzweilin, same SHA as #23's head — not rebased, not cherry-picked, not squashed into anything of mine. My five commits sit on top.

The base branch was only ever a way to make the diff show my delta alone; it cost the commit list showing who wrote what, which is the worse trade.

@mzweilin — you offered to close #23 and take this one instead. If that is the call, please use **Rebase and merge**: merge commits are disabled on this repository and a squash would land your workflow as one commit authored by me. Landing #23 first is equally fine and I will rebase this onto `main` after it.

PR #26, opened 2026-09-16
chore: move the intel/gpu-ai-skills pin to 0b4fafd09c5e

xaleryb:
`intel/gpu-ai-skills` has moved on `main` and this repository's copy of it has not. Opened by `tools/check_upstream.py`; the bytes are upstream's at the new commit, written by `tools/sync_external.py --write`.

- pin: `3fb51a0b69527fadcdec3e4ae6b84121d39115dc` -> `0b4fafd09c5eb4cc5daf532d915ef5984a919775`
- upstream diff: https://github.com/intel/gpu-ai-skills/compare/3fb51a0b6952...0b4fafd09c5e
- changed upstream: `model-config-recommend`, `torch-xpu-run`, `vllm-xpu-bench`, `vllm-xpu-profile`, `vllm-xpu-run`, `xpu-container-run`, `xpu-deploy-plan`, `xpu-profile-unitrace`, `xpu-system-setup`
- unchanged upstream, re-vendored because the pin moved: `cuda-to-xpu-migration`, `llamacpp-xpu-run`, `model-can-it-fit`, `sglang-xpu-bench`, `sglang-xpu-run`, `torch-xpu-bench`, `torch-xpu-profile`, `xpu-discover`, `xpu-model-type-detect`, `xpu-port`, `xpu-runtime-preflight`

What this does not decide is whether the new text should be merged. An import is somebody else's document, so a change in it is a change somebody made there: the checks below answer the structural half, and the diff is worth reading.

If no checks are listed, this pull request was opened with `GITHUB_TOKEN`, which does not start workflow runs. Push to the branch, or close and reopen it, to get them.

PR #27, opened 2026-09-16
feat: propose the pin move when an imported skill changes upstream

xaleryb:
`sync_external.py --check` proves an imported skill is still the bytes of its pinned commit. Nothing proved the pin is still the commit upstream maintains, and that failure is silent: every check here stays green while what this catalog publishes falls behind what its authors ship. This is the other half.

- **`tools/check_upstream.py`** — compares the tree object id of each `external-path` at the pin against the same path at the tip of upstream's default branch. When a pinned directory moved, it moves `external-commit` everywhere this repository writes it down (every entry of that upstream, the twelve characters the group comment quotes, and `NOTICE`), re-vendors with `sync_external.py --write`, and opens one pull request per upstream on `sync/<upstream>-<commit>`.
- **`tools/upstream_git.py`** — the transport, now shared with `sync_external.py`.
- **`.github/workflows/upstream-sync.yml`** — weekly plus `workflow_dispatch` with a `dry-run` input; runs only in `intel/skills`; the one workflow here that writes.
- **`validate.yml`** — the detector's self-test, beside the two self-tests already there, because a scheduled job that stops detecting is not noticed until a Monday.

**What it does not do** is decide whether the new text should be published. An import is somebody else's document, so a change in it is a change somebody made there: the pull request's checks answer the structural half and a reviewer reads the diff.

**Why a directory hash and not upstream's HEAD.** Most commits upstream touch none of the skills imported here, so following HEAD would open a pull request that moves 23 pins, rewrites 23 `.source.json` files and changes no skill text. Per-path comparison costs 13.75 KiB of transfer per upstream and the whole survey runs in about three seconds; nothing is checked out, so the same code runs offline as the self-test above.

**No third-party code receives the write token** — plain `git` and `gh`, the stance `security.yml` already states.

**Token.** GitHub starts no workflow run for anything `GITHUB_TOKEN` does, including the branch push, so a pull request opened with it arrives with no checks. The job takes `UPSTREAM_SYNC_TOKEN || GITHUB_TOKEN`, warns in the run when it falls back, and every pull request it opens says which case it is in its body.

## Run, not assumed

- **#26 is this tool's own output**, opened by it against this repository: `intel/gpu-ai-skills` moved `3fb51a0b6952 -> 0b4fafd09c5e`, 9 of 20 pinned directories changed, 37 files, and the level 1 gate is green on it. `intel/intel-performance-skills` is at its pin and was correctly left alone.
- **`--self-test`**: 21 assertions against this tree's real pins, offline. `--mutate M1..M9` is nine distinct ways the detector can break — every one of them reports "nothing moved upstream", and every one of them turns the self-test red.
- **Idempotence on the live pull request**: a second run answers `SKIP intel/gpu-ai-skills: #26 (open) already proposes sync/gpu-ai-skills-0b4fafd09c5e`, including the case where the head branch lives in a fork.
- **Failure contracts**: an uncommitted working tree fails rather than carrying its changes into a pull request; a pin that no longer resolves fails; an upstream that cannot be reached warns and exits 0, for the same reason the link check does.
- `actionlint` and `zizmor --min-severity=medium --min-confidence=medium` are clean.

**Not run yet:** the workflow itself has never executed on a runner — the schedule, the empty `dry-run` input on a `schedule` event, `env.SYNC_TOKEN_SET` and the repository guard are static-checked only. A `workflow_dispatch` with `dry-run: true` right after this merges is what closes that, and it changes nothing when it runs.

Docs are in the same change: `MAINTAINERS.md` gains the workflow row, what the write permissions are for, the token trap, and how to run the same proposal by hand from a fork.


napetrov (approved):

vchamarthi:
Non-blocking: detection is per-directory, but bump() re-vendors every skill in the upstream group.
On #26, all 20 .source.json files were rewritten but only 9 of 20 skills changed. changed_skills.py saw 20, so skillspector ran 20 scans ([run 35152850182](https://github.com/intel/skills/actions/runs/35152850182) (https://github.com/intel/skills/actions/runs/35152850182)), 11 of them over identical bytes.

classify() already knows which dirs changed. Follow-up: 
changed_skills.py could skip skills whose only diff is .source.json. The weekly cron still catches scanner-rule drift.

